PR #1, opened 2019-05-23
Feature/co1442 open api design

nosnatef:


README.md
```diff
@@ -1,318 +1,336 @@
...
+            format: url
+            description: A link to further information about the error
+            example: https://developer.oregonstate.edu/documentation/error-reference#1234
+  Errors:
+    properties:
+      errors:
+        type: array
+        items:
+          $ref: '#/definitions/Error'
+
```
tsoliangwu0130: I think you pasted your openapi design to the readme accidentally, can you fix it?

openapi.yaml
```diff
@@ -16,175 +16,148 @@ host: api.oregonstate.edu
...
-  url: https://github.com/osu-mist/express-api-skeleton
+  url: https://github.com/osu-mist/poker-api
 produces:
   - application/json
 paths:
-  /pets:
-    get:
-      summary: List all pets
+  /games:
+    post:
```
tsoliangwu0130, line 24: Currently, you only have a method for your API: `POST /games`, is it correct?

tsoliangwu0130 (changes requested):

openapi.yaml
nosnatef: I added more methods. Does that look good?

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+      summary: Create a poker game
       tags:
-        - pets
-      description: Get all pets
-      operationId: getPets
+        - games
+      consumes:
+        - application/json
+      operationId: postGames
```
tsoliangwu0130: Since this method will only create one game object at a time, I think the appropriate `operationId` should be `postGame`.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
       parameters:
+        - in: body
+          name: requestBody
+          schema:
+            $ref: '#/definitions/GamePostBody'
+      responses:
+        '201':
+          description: Successfully update a game
```
tsoliangwu0130: Should be `create` when using POST method.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+              format: url
+              description: Location of the newly created game
+        '400':
+          description: Invalid game object
+          schema:
+            $ref: '#/definitions/Errors'
+        '404':
+          description: No such a game
+          schema:
+            $ref: '#/definitions/Errors'
```
tsoliangwu0130, line 54: When will you get a `404` when sending a POST request to this endpoint?

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+  /games/{gameID}:
```
tsoliangwu0130: Since the OpenAPI file will be used as a single source of truth, we also follow this [JS naming convention](https://github.com/osu-mist/style-guides/blob/master/JavaScript.md#naming-conventions) in our style guild to avoid the inconsistency. Based on it, we should us `Id` instead of `ID`. Can you check through your OpenAPI file and fix them all?

nosnatef:
All methods in README.md are added in the openapi.yaml

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
-        - in: query
-          name: species
-          description: Search by pet species
-          required: false
-          type: string
+        - name: gameID
+          in: path
+          description: unique integer
+          required: true
+          type: integer
```
tsoliangwu0130: Based on JSON API, object ID should always be a string: https://jsonapi.org/format/#document-resource-object-identification
Example: https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L893-L896

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+            $ref: '#/definitions/GameResult'
+        '404':
+          description: Game not found
+          schema:
+            $ref: '#/definitions/Errors'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+    put:
```
tsoliangwu0130: Why using PUT for updating? Take a look at this paragraph: https://jsonapi.org/faq/#wheres-put

That said, we should use `PATCH` for updating resource especially your update method allow partially update.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+      tags:
+        - games
+      summary: Update a game specified by its id
+      operationId: updateGameByID
+      consumes:
+        - application/json
+      produces:
+        - application/json
```
tsoliangwu0130: Didn't you specify the `produces` type at very top-level [here](https://github.com/osu-mist/poker-api/pull/1/files#diff-fe030a7c1568b3decf599edf399be7f4R20)?

> An API can accept and return data in different formats, the most common being JSON and XML. You can use the consumes and produces keywords to specify the MIME types understood by your API. The value of consumes and produces is an array of MIME types. Global MIME types can be defined on the root level of an API specification and are inherited by all API operations.

Ref: https://swagger.io/docs/specification/2-0/mime-types/

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+      parameters: 
+        - name: gameID
+          in: path
+          description: ID of game to be updated
+          required: true
+          type: integer
```
tsoliangwu0130: So, the `gameID` was actually used multiple times for various methods. Do you think we should add it under parameters and make it reusable? For example: https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L893-L896

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+        - name: updateBody
+          in: body
+          description: Object that will overwrite the game specified by its id
+          required: true
+          schema:
+            $ref: '#/definitions/GameUpdateBody'
+      responses:
+        '200':
+          description: Successfully update a Game
```
tsoliangwu0130: typo: Game -> game

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+  /games/{gameID}:
+    get:
+      summary: Info for a specific game
+      tags:
+        - games
+      description: get game by game id
```
tsoliangwu0130: grammar mistake: get game by game id -> Get a game by its gameId

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+      responses:
+        '200':
+          description: Successfully update a Game
+          schema:
+            $ref: '#/definitions/GameResult'
+          headers:
+            Location:
+              type: string
+              format: url
+              description: Location of the updated game
```
tsoliangwu0130: For successful updating, I don't you'll need to return header with the new self link: https://jsonapi.org/format/#crud-updating-responses-200

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+          description: Game not found
+          schema:
+            $ref: '#/definitions/Errors'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+  /players/{playerID}:
+    get:
+      summary: Info for a specific player in a game
```
tsoliangwu0130: By your definition, the player should be globally unique, so the summary for this endpoint is not accurate since a player is not necessary to be in a game.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+          pattern: '^g[1-9]\d*p[1-9]\d*$'
       responses:
         '200':
           description: Successful response
           schema:
-            $ref: '#/definitions/PetResult'
+            $ref: '#/definitions/PlayerResult'
         '404':
-          description: PetResource not found
+          description: Game not found
```
tsoliangwu0130: Should be player not found.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+  /games/{gameID}/players:
     get:
-      summary: Info for a specific pet
+      summary: Get a list of players in the game
       tags:
-        - pets
-      description: Get pet by unique ID
-      operationId: getPetById
-      parameters:
+        - players
```
tsoliangwu0130, line 152: Since it is also a method under the `/games` endpoint so I think it'll make more sense to add a `games` tag as well.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+          schema:
+            $ref: '#/definitions/Errors'
+    put:
+      tags:
+        - players
+      summary: Update a player specified by its id
+      operationId: updatePlayerByID
+      consumes:
+        - application/json
+      produces:
```
tsoliangwu0130: Again, you've defined the `produces` type before.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+        - application/json
+      parameters: 
+        - name: playerID
```
tsoliangwu0130: Same comment, `playerID` was used multiple times, so it'll be better to define it under parameters to make it reusable.

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+          description: Object that will overwrite the player specified by its id
+          required: true
+          schema:
+            $ref: '#/definitions/PlayerUpdateBody'
+      responses:
+        '200':
+          description: Successfully update a Player
+          schema:
+            $ref: '#/definitions/PlayerResult'
+          headers:
```
tsoliangwu0130, line 324: Same, it's no need to return a headers for successful updating based on JSON API standard.

openapi.yaml
```diff
@@ -82,109 +222,252 @@ parameters:
...
-  pageSize:
-    name: page[size]
-    in: query
-    type: integer
-    required: false
-    description: Number of results to return
-    default: 25
-    maximum: 500
-    minimum: 1
+    description: '"Bearer [token]" where token is your OAuth2 access token'
```
tsoliangwu0130, line 420: What if you tried to update a player hasn't created yet?

openapi.yaml
```diff
@@ -16,60 +16,200 @@ host: api.oregonstate.edu
...
+              description: Location of the updated game
         '400':
-          description: Bad request
+          description: Invalid game object
           schema:
             $ref: '#/definitions/Errors'
         '500':
           description: Internal server error
           schema:
             $ref: '#/definitions/Errors'
```
tsoliangwu0130, line 129: What if you tried to update a game hasn't created?

openapi.yaml
```diff
@@ -82,109 +222,252 @@ parameters:
...
+      $ref: '#/definitions/player'
+  player:
+    type: object
+    properties:
+      cards:
+        $ref: '#/definitions/cards'
+      bet:
+        $ref: '#/definitions/playerBet'
+      status:
+        $ref: '#/definitions/playerStatus'
```
tsoliangwu0130: How can you tell which play is who? For example, here is the example response your clients will get from `GET /games/{gameID}`:

```json
{
  "links": {
    "self": "string"
  },
  "data": {
    "id": 0,
    "type": "game",
    "attributes": {
      "round": "blind",
      "players": [
        {
          "cards": [
            {
              "attributes": {
                "cardNumber": "A",
                "cardSuit": "Diamond"
              }
            }
          ],
          "bet": 0,
          "status": "called"
        }
      ],
      "tableCards": [
        {
          "attributes": {
            "cardNumber": "A",
            "cardSuit": "Diamond"
          }
        }
      ]
    },
    "links": {
      "self": "string"
    }
  }
}
```

Inside of the `players` array, since we won't know the player ID or the self-link of the player, how can we call POST/PUT method to update their info?

openapi.yaml
```diff
@@ -82,109 +222,252 @@ parameters:
...
+    type: integer
+    minimum: 0
+    description: Each player's bet
+  playerStatus:
+    type: string
+    enum:
+      - "called"
+      - "raised"
+      - "checked"
+      - "folded"
```
tsoliangwu0130: No need to have quotes for these strings: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

openapi.yaml
```diff
@@ -82,109 +222,252 @@ parameters:
...
-  PetResult:
+          players:
+            $ref: '#/definitions/players'
+  card:
+    properties:
+      attributes:
+        type: object
+        properties:
+          cardNumber:
+            $ref: '#/definitions/cardNumber'
```
tsoliangwu0130: Actually, `cardNumber` is only used by this object. Here are a few things we should think about carefully when writing code (no matter in what languages):

1. readability
2. reusability
3. extensibility
4. maintainability

Just my two cents, I think there are no obvious reasons to separate these definitions to different objects and it actually increases the difficulty of readability of your code. For those only used one time and straight forward, we usually directly define it inside where it should be. For example: For example: https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L820-L863

Same comments with other definitions below.

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
-        - pets
-      description: Get pet by unique ID
-      operationId: getPetById
+      - games
+      summary: "Remove a game specified by its id"
+      operationId: removeGameByID
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
```
tsoliangwu0130: Does a delete method require any JSON input/output?

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+            required:
+              - round
+              - players
+              - tableCards
+            additionalProperties: false
+        required: 
+          - type
+          - attributes
+        additionalProperties: false
+  PlayerUpdateBody:
```
tsoliangwu0130: Based on JSON API: https://jsonapi.org/format/#crud-updating, the PATCH request MUST include a single resource object as primary data. The resource object MUST contain `type` and `id` members.

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+    type: integer
+    minimum: 0
+    description: The minimum bet every player have to place at least each round.
+    example: 1000
+  maximumBet:
+    type: integer
+    minimum: 0
+    description: The maximum bet every player can place each round.
+    example: 2000
+  playerLinks:
```
tsoliangwu0130: Where did you use this object?

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+    enum:
+      - game
+  playerType:
+    type: string
+    enum:
+      - player
+  playerID:
+    type: string
+    example: 'g1p1'
+    pattern: '^g[1-9]\d*p[1-9]\d*$'
```
tsoliangwu0130: So here is my question. Should a player be globally unique? If yes, why the `playerId` should start with the `gameId`? If not, why we need to provide a global `GET /players/{playerId}` method? Shouldn't we just make the method under `/games` like this `GET /games/{gameId}/players/{playerId}`?

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+              description: Location of the updated player
+        '400':
+          description: Invalid player object
           schema:
             $ref: '#/definitions/Errors'
         '500':
           description: Internal server error
           schema:
             $ref: '#/definitions/Errors'
+    delete:
```
tsoliangwu0130, line 338: Just curious, why we need a DELETE method for `/players` endpoint? Based on your design, we [could update the players list for each game](https://github.com/osu-mist/poker-api/blob/feature/CO1442-OpenAPI-design/openapi.yaml#L406). Having this method won't hurt, but I'm wondering in what scenario this DELETE will be called?

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+    description: The number of player in the game
+    example: 2
+  minimumBet:
+    type: integer
+    minimum: 0
+    description: The minimum bet every player have to place at least each round.
+    example: 1000
+  maximumBet:
+    type: integer
+    minimum: 0
```
tsoliangwu0130: Can `maximumBet` ever be `0`?

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+      $ref: '#/definitions/playerLink'
+  playerLink:
+    type: string
+    format: url
+    description: Link to the playerResource
+  cards:
+    type: array
+    items:
+      $ref: '#/definitions/card'
+  cardLink:
```
tsoliangwu0130: What is `cardLink`? Do you have card resource?

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+              round:
+                $ref: '#/definitions/round'
+              players:
+                $ref: '#/definitions/players'
+              tableCards:
+                $ref: '#/definitions/cards'
+            required:
+              - round
+              - players
+              - tableCards
```
tsoliangwu0130: When will the game update be called? Before the game start or during the game or both? Here are some scenarios I can think of:

1. After the game created and before the game starts: players are possible to join to and leave from a game, so we can call this function to update the players list. Is it required to contain `round` or `tableCards` in the data body at this stage? Also, should we allow users to modify min/max bet and players number at this stage?
2. During the game, the players list shouldn't be modified which means the `players` field shouldn't be required here. However, we can still update the game round and table cards by using this method.

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+          description: g[int]p[int] where the first int is the ID of the game and the second int is the ID of the player in that game.
           required: true
           type: string
+          pattern: '^g[1-9]\d*p[1-9]\d*$'
       responses:
         '200':
           description: Successful response
           schema:
-            $ref: '#/definitions/PetResult'
+            $ref: '#/definitions/PlayerResult'
```
tsoliangwu0130, line 220: Based on your definition, here is an example of the returned PlayerResult:

```json
{
  "links": {
    "self": "string"
  },
  "data": {
    "id": "g1p1",
    "type": "player",
    "attributes": {
      "players": [
        {
          "cards": [
            {
              "attributes": {
                "cardNumber": "A",
                "cardSuit": "Diamond"
              }
            }
          ],
          "bet": 0,
          "status": "called"
        }
      ]
    }
  }
}
```

Should the `players` array be included in a single player object?

README.md
```diff
@@ -1,318 +1,55 @@
...
-                 -p 8080:8080 \
-                 -p 8081:8081 \
-                 -v path/to/keytools/:/usr/src/<my-api>/keytools:ro \
-                 -v "$PWD"/config:/usr/src/<my-api>/config:ro \
-                 -v "$PWD"/logs:/usr/src/<my-api>/logs \
-                 --name <my-api> \
-                 <my-api>
-    ```
+ ### /players/{id}
+Delete a player from a certain game(player leaving the game).
```
tsoliangwu0130: Since we've used OpenAPI/Swagger file to define the API specification and behavior, there is no need to repeat them again in your README. However, we do wanna include how to install, run and test all kinds of these instruction steps on our README file. You can still keep the description and purpose parts in this file, but I'd suggest pulling the rest part from the skeleton again and change it as needed.

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+          description: "Internal server error"
+          schema:
+            $ref: "#/definitions/Error"
+  /games/{gameID}/players:
+    get:
+      summary: Get a list of players in the game
+      tags:
+        - players
+      description: get players by game id
+      operationId: getPlayersByGameID
```
tsoliangwu0130: Nice, it's clear. Since you will two different kinds of IDs in your API (`GameId` and `PlayerId`), I'd suggest renaming the rest of operationId to specify which ID is using for the current method. For example:

1. `getPlayerByID` -> `getPlayerByPlayerId`
2. `removeGameByID` -> `removeGameByGameId`

and so on.

tsoliangwu0130 (changes requested):

openapi.yaml
tsoliangwu0130: In addition, how can we create a player?

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+            $ref: '#/definitions/Errors'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+    put:
+      tags:
+        - games
+      summary: Update a game specified by its id
+      operationId: updateGameByID
```
tsoliangwu0130: I think we should use the same HTTP verbs as you did for `operationId: postGames` for the consistency. Same comments with your other methods.

openapi.yaml
```diff
@@ -16,175 +16,511 @@ host: api.oregonstate.edu
...
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - name: playerID
+          in: "path"
+          description: "Id of player to be removed"
+          required: true
+          type: string
+          pattern: '^g[1-9]\d*p[1-9]\d*$'
```
tsoliangwu0130: Since you've defined there can be [at most 5 players](https://github.com/osu-mist/poker-api/pull/1/files#diff-fe030a7c1568b3decf599edf399be7f4R459) in a game, will `g3p6` or `g3p10` be a valid playerId?

openapi.yaml
nosnatef: The player will only exist if a game exist. It is globally unique as it is summoned and before it is deleted as the game ends.

openapi.yaml
nosnatef: It would not be allowed.

openapi.yaml
nosnatef: It would not be allowed.

openapi.yaml
nosnatef: The player will be identified as g1p1, g1p2.... etc. The self link will be in links.self

openapi.yaml
nosnatef: The player is globally unique as you can access its information using gxpx. The player will not be needed after the game.

nosnatef: In my design there is no need to use a player twice, but as long as it lives through a lifespan of game, it's still globally unique.

openapi.yaml
nosnatef: The player will be created upon the creation of game.

openapi.yaml
nosnatef: It shouldn't, but I am not sure how small it can be since it can be really, really small (like .0000000000000001)

nosnatef:
The player is created upon the creation of the game. You can understand player as a human with only 5 minute to live. It does exist in this world, you can talk to them, find out what they look like but they will die in 5 minutes. I am not sure if this is a good way to explain my design.

openapi.yaml
tsoliangwu0130: What do you mean "not allowed"? What status code will be returned in this case?

openapi.yaml
tsoliangwu0130: `/players/{playerID}` sounds more like trying to get a player's information from the database, no matter it belongs to a game or not. Based on what you described, the endpoint should look like `/games/{gameID}/players/{playerID}`, and the `playerID` won't need to be globally unique then; it only to be unique among one games (In other words, game 1 and game 2 can both has player 1, but they can be different persons).

openapi.yaml
tsoliangwu0130: But how? Do you have any endpoint for this operation?

nosnatef:
Ian and I are talking about if we can change paths like GET /players/{playerId} to GET /games/{gameId}/players/{playerId}(The player that is bond to the game, will be cleaned after that).
He is not sure if this is allowed of OpenAPI specification.

openapi.yaml
tsoliangwu0130: Again, what does "not allowed" mean? You need to define a proper response with its status code for all possible scenarios so that your clients can expect what they will receive when using your API.

openapi.yaml
tsoliangwu0130: Where will the `playerID` go? From the response schema I attached, there is no `playerID` field in your players/player object.

openapi.yaml
nosnatef: See my new comment

openapi.yaml
tsoliangwu0130: `.0000000000000001` is not an integer.

nosnatef: I think i will change it to 1

README.md
tsoliangwu0130: Did you address this?

openapi.yaml
```diff
@@ -215,3 +414,294 @@ definitions:
...
+      - data
+    additionalProperties: false
+  Round:
+    type: string
+    enum:
+      - 'blind'
+      - 'flop'
+      - 'turn'
+      - 'river'
+      - 'showdown'
```
tsoliangwu0130: Quotes are not needed for these enums.

openapi.yaml
```diff
@@ -16,175 +16,374 @@ host: api.oregonstate.edu
...
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+  /games/{gameId}:
+    get:
+      summary: Info for a specific game
+      tags:
+        - games
+      description: Get a game by its gameid
```
tsoliangwu0130, line 80: typo: `gameid` -> `gameId`

openapi.yaml
```diff
@@ -16,175 +16,374 @@ host: api.oregonstate.edu
...
           schema:
             $ref: '#/definitions/Errors'
         '500':
           description: Internal server error
           schema:
             $ref: '#/definitions/Errors'
-  /pets/{id}:
+    delete:
+      tags:
+      - games
```
tsoliangwu0130: Indentation is missing.

openapi.yaml
```diff
@@ -16,175 +16,374 @@ host: api.oregonstate.edu
...
+        - $ref: '#/parameters/gameId'
+      responses:
+        204:
+          description: "game successfully removed"
+        404:
+          description: "No game with this id"
+          schema:
+            $ref: "#/definitions/Error"
+        500:
+          description: "Internal server error"
```
tsoliangwu0130: Quotes are not necessary for all descriptions fields but are needed for the status codes. I'd recommend go through the entire OpenAPI file again to make sure it follows our style guild.

openapi.yaml
```diff
@@ -16,175 +16,374 @@ host: api.oregonstate.edu
...
+          schema:
+            $ref: "#/definitions/Error"
+  /games/{gameId}/players:
     get:
-      summary: Info for a specific pet
+      summary: Get a list of players in the game
+      tags:
+        - games
+      description: get players by game id
+      operationId: getPlayersBygameId
```
tsoliangwu0130: typo: `getPlayersBygameId` -> `getPlayersByGameId`

openapi.yaml
```diff
@@ -16,60 +16,305 @@ host: api.oregonstate.edu
...
       parameters:
+        - $ref: '#/parameters/gameId'
+      responses:
+        204:
+          description: game successfully removed
+        404:
+          description: No game with this id
+          schema:
+            $ref: "#/definitions/Error"
+        500:
```
tsoliangwu0130: These response codes need to be quoted since we're expecting strings but they could be parsed as numbers: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

openapi.yaml
```diff
@@ -82,109 +327,58 @@ parameters:
...
+                - 4
+                - 5
+                - 6
+                - 7
+                - 8
+                - 9
+                - 10
+                - J
+                - Q
+                - K
```
tsoliangwu0130: In this enum, 2 - 10 could be parsed as numbers so they should be quoted to avoid the confusion. In addition, since 2 - 10 will need to be quoted, I'm fine with quoting `A`, `J`, `Q`, `K` as well because of the consistency, but it's up to you.

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+              PlayerIds:
+                $ref: '#/definitions/PlayerIds'
+              MinimumBet:
+                $ref: '#/definitions/MinimumBet'
+              MaximumBet:
+                $ref: '#/definitions/MaximumBet'
+              BetPool:
+                $ref: '#/definitions/BetPool'
+              TableCards:
+                $ref: '#/definitions/Cards'
```
tsoliangwu0130, line 531: As I mentioned in this [comment](https://github.com/osu-mist/poker-api/pull/1#discussion_r287173942), we only want to separate the definitions if needed (e.g. the definition can be reused or there is a specific reason like [this](https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#result-object--resource-object)). Can you adjust them to make the code easier to read a bit?

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+                $ref: '#/definitions/Round'
+              PlayerIds:
+                $ref: '#/definitions/PlayerIds'
+              MinimumBet:
+                $ref: '#/definitions/MinimumBet'
+              MaximumBet:
+                $ref: '#/definitions/MaximumBet'
+              BetPool:
+                $ref: '#/definitions/BetPool'
+              TableCards:
```
tsoliangwu0130: The attribute names here are all start with upper case letter. The attribute names should always follow our naming convention [here](https://github.com/osu-mist/style-guides/blob/master/JavaScript.md#naming-conventions). Can you check through all of your other definition to make sure they adhere to the standard?

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+    minimum: 1
+    description: The minimum bet every player have to place at least each round.
+    example: 1000
+  PlayerBet:
+    type: integer
+    minimum: 0
+    description: The bet from a specific player.
+  PlayerEmail:
+    type: string
+    example: 'abc@efg.com'
```
tsoliangwu0130: Not need to quote this, also, you can add a `format: email` to this definition. Ref: https://swagger.io/docs/specification/data-models/data-types/ (see the `String Formats` section).

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+      data:
+        $ref: '#/definitions/PlayerInGameResource'
+  PlayerLevel:
+    type: integer
+    minimum: 1
+    description: Player's level in the system.
+    example: 20
+  PlayerName:
+    type: string
+    example: 'Jack'
```
tsoliangwu0130: No need to quote.

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+  PlayerPassword:
+    type: string
+    example:  'hunter2'
```
tsoliangwu0130: No need to quote. Also, you can add `format: password` to this definition as well.

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+            additionalProperties: false
+        required: 
+          - type
+          - attributes
+        additionalProperties: false
+    required: 
+      - data
+    additionalProperties: false
+          
+
```
tsoliangwu0130: Not necessary blank lines.

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+    type: integer
+    minimum: 0
+    description: Player's experience point above current level.
+    example: 1234
+  PlayerId:
+    type: string
+    example: '101'
+  PlayerIds:
+    properties:
+      players:
```
tsoliangwu0130: Your current `playerIds` object will look like this:

```
{
    playerIds: {
        players: ['101', '102']
    }
}
```

Why do you need a nested `players` attribute inside of the `playerIds`? I guess you should be able to store the array directly under `playerIds` like this:

```
{
    playerIds: ['101', '102']
}
```

openapi.yaml
```diff
@@ -82,109 +327,58 @@ parameters:
...
-      data:
-        $ref: '#/definitions/PetResource'
-  PetsResult:
+  BetPool:
+    type: integer
+    minimum: 0
+    description: Total bet from players.
+  Card:
+      properties:
+        attributes:
```
tsoliangwu0130: Do we need to wrap the `cardNumber` and `cardSuit` in `attributes`?

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+          BetPool:
+            $ref: '#/definitions/BetPool'
+          TableCards:
+            $ref: '#/definitions/Cards'
+      links:
+        $ref: '#/definitions/SelfLink'
+  GameResult:
+    properties:
+      data:
+        $ref: '#/definitions/GameResource'
```
tsoliangwu0130, line 594: Here is the example response based on your current design:

```
{
  "games": [
    {
      "data": {
        "id": "string",
        "type": "game",
        "attributes": {
          "Round": "blind",
          "PlayerIds": {
            "players": [
              "101"
            ]
          },
          "MinimumBet": 1000,
          "MaximumBet": 2000,
          "BetPool": 0,
          "TableCards": {
            "cards": [
              {
                "attributes": {
                  "CardNumber": "A",
                  "CardSuit": "Diamond"
                }
              }
            ]
          }
        },
        "links": {
          "self": "string"
        }
      }
    }
  ]
}
```

As you can see, the top-level attribute is `games` and it's not allowed. According to [JSON API spcification](https://jsonapi.org/format/#document-top-level), the top-level members *MUST* contains one of these:

* `data`: the document’s “primary data”
* `errors`: an array of error objects
* `meta`: a meta object that contains non-standard meta-information.

You're now actually wrapping your game object under the `data` attribute in each element of the array, which is not correct. Take a look at the document and modify your response to adhere to the standard. You can also take a look at our other APIs to see how we usually define the response object for various endpoints. For example: https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L300-L307

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
+          PlayerLevel:
+            $ref: '#/definitions/PlayerLevel'
+          PlayerExpAboveLevel:
+            $ref: '#/definitions/PlayerExpAboveLevel'
+          PlayerBet:
+            $ref: '#/definitions/PlayerBet'
+          PlayerStatus:
+            $ref: '#/definitions/PlayerStatus'
+          PlayerCards:
+            $ref: '#/definitions/Cards'
```
tsoliangwu0130, line 688: Nice! It's not necessary but I'd recommend also add the current game number in this response to make it more clear.

openapi.yaml
tsoliangwu0130: This comment also applied for all of your multi-resource endpoints, e.g. `GET /players`.

openapi.yaml
```diff
@@ -215,3 +409,314 @@ definitions:
...
           $ref: '#/definitions/Error'
+  GameId:
+    type: string
+    description: A unique ID of a poker game
+  GameIds:
+    properties:
+      games:
+        type: array
+        items:
+          $ref: '#/definitions/GameId'
```
tsoliangwu0130: Same comments as above. You don't need to wrap the array under `games` object.

tsoliangwu0130 (changes requested):
A few more comments:

1. Can you add the functionality to filter result to your API? For example, if I want to search for the games with the round is blind, how can you do that? Feel free to add as many filters as you want to all proper endpoints.
2. Do you think we should have an endpoint of `GET /games/gameId/players`? This endpoint could return an array of players object (not only the player IDs, but all detailed gaming related information).
3. Since the `GET /players/playerId` and `GET /games/gameId/players/playersId` are returning two different resources, to be more clear, I'd suggest to rename `/players/playerId` to `/members/memberId`. And the `playerId` from `GET /games/gameId/players/playersId` should be the competition with `gameId` and `memberId`. Feel free to let me know if you have any questions on this, we can talk about them in person.

openapi.yaml
nosnatef: BetPool is reused in both GameResult and GamePatchBody

openapi.yaml
nosnatef: I think we do.

nosnatef:
Waiting for further code review.

openapi.yaml
```diff
@@ -16,176 +16,405 @@ host: api.oregonstate.edu
...
-      prev:
-        type: string
-        format: url
-        description: The previous page of data
-      next:
-        type: string
-        format: url
-        description: The next page of data
-  PetResource:
+  betPool:
```
tsoliangwu0130: Definition object should use `UpperCamelCase ` like the class naming convention.

openapi.yaml
```diff
@@ -16,176 +16,405 @@ host: api.oregonstate.edu
...
+          - '5'
+          - '6'
+          - '7'
+          - '8'
+          - '9'
+          - '10'
+          - 'J'
+          - 'Q'
+          - 'K'
+      CardSuit:
```
tsoliangwu0130: The JSON field name should be `normalCamelCase` (start with lower case).

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+          playerBet:
+            $ref: '#/definitions/playerBet'
+          playerStatus:
+            $ref: '#/definitions/playerStatus'
+          playerCards:
+            $ref: '#/definitions/cards'
+  playerInGameResult:
+    properties:
+      data:
+        $ref: '#/definitions/playerInGameResource'
```
tsoliangwu0130: Here is an example player result based on your design:

```{
  "data": {
    "id": "101",
    "type": "player",
    "attributes": {
      "memberName": "Jack",
      "memberLevel": 20,
      "memberExpAboveLevel": 1234,
      "playerBet": 0,
      "playerStatus": "folded",
      "playerCards": {
        "cards": [
          {
            "cardNumber": "A",
            "CardSuit": "Diamond"
          }
        ]
      }
    }
  }
}
```

It's no need to put a `cards` array inside of `playerCards`, instead, you can directly make `playerCards` become an array. Same with `tableCards` in `gameResult`, `PATCH`, and `POST` body.

tsoliangwu0130 (changes requested):

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+              betPool:
+                $ref: '#/definitions/betPool'
+              tableCards:
+                $ref: '#/definitions/cards'
+            required:
+              - id
+            additionalProperties: false
+        required: 
+          - type
+          - attributes
```
ianbrown9475: From the jsonapi spec about updating resources https://jsonapi.org/format/#crud-updating:
> The resource object MUST contain type and id members.

> Any or all of a resource’s attributes MAY be included in the resource object included in a PATCH request.

So it looks like `type` and `id` should be required and `attributes` should not be.

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+              minimumBet:
+                $ref: '#/definitions/minimumBet'
+              maximumBet:
+                $ref: '#/definitions/maximumBet'
+              betPool:
+                $ref: '#/definitions/betPool'
+              tableCards:
+                $ref: '#/definitions/cards'
+            required:
+              - id
```
ianbrown9475: This `required` is referencing the `attributes` object which doesn't have a property `id`. I think this should be referencing the `data` object instead

openapi.yaml
ianbrown9475: Like @tsoliangwu0130 mentioned, `- games` should be indented here to be consistent with the rest of the document

openapi.yaml
```diff
@@ -16,176 +16,405 @@ host: api.oregonstate.edu
...
+        - $ref: '#/parameters/authorization'
+        - $ref: '#/parameters/playerId'
+        - $ref: '#/parameters/gameId'
+      responses:
+        '200':
+          description: Successful response
+          schema:
+            $ref: '#/definitions/playerInGameResult'
+        '404':
+          description: Player not found
```
ianbrown9475: Since a 404 could also mean that the game was not found, you should specify this in the description

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+  gameId:
+    type: string
+    description: A unique ID of a poker game
+  gameIds:
+    type: array
+    items:
+      $ref: '#/definitions/gameId'
+  gamePatchBody:
+    properties:
+      data:
```
ianbrown9475, line 512: `data` should also be required and `additionalProperties` should be `false`

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+              playerBet:
+                $ref: '#/definitions/playerBet'
+              playerStatus:
+                $ref: '#/definitions/playerStatus'
+              playerCards:
+                $ref: '#/definitions/cards'
+            additionalProperties: false
+            required:
+              - id
+              - type
```
ianbrown9475: `required` and its items need to be unindented so that it's under the `data` object

openapi.yaml
```diff
@@ -16,176 +16,405 @@ host: api.oregonstate.edu
...
+        '200':
+          description: Successfully update a player from game
+          schema:
+            $ref: '#/definitions/errors'
+        '400':
+          description: Invalid player in game object
+          schema:
+            $ref: '#/definitions/errors'
+        '404':
+          description: Player not found.
```
ianbrown9475: Same comment here about how a 404 could mean a game is not found also

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+        additionalProperties: false
+    required: 
+      - data
+    additionalProperties: false 
+  memberResource:
+    properties:
+      id:
+        $ref: '#/definitions/playerId'
+      type:
+        $ref: '#/definitions/playerType'
```
ianbrown9475: The `member` resource can't have the same `type` as the `player` resource. Otherwise, there could exist two resources with the same `type` and same `id` which would violate the jsonapi spec. You will need to have a different `type` if you want to use the same `id`.

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+            $ref: '#/definitions/playerType'
+          attributes:
+            type: object
+            properties:
+              memberName:
+                $ref: '#/definitions/memberName'
+              playerEmail:
+                $ref: '#/definitions/playerEmail'
+              playerPassword:
+                $ref: '#/definitions/playerPassword'
```
ianbrown9475: These fields in the post body should match the field names in the `member` resource

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+  memberName:
+    type: string
+    example: Jack
+  playerPassword:
+    type: string
+    example:  hunter2
+    format: password
+  memberPatchBody:
+    properties:
+      data:
```
ianbrown9475, line 745: `data` should be required and `additionalProperties` should be false

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+              playerEmail:
+                $ref: '#/definitions/playerEmail'
+              playerPassword:
+                $ref: '#/definitions/playerPassword'
+              memberLevel:
+                $ref: '#/definitions/memberLevel'
+              memberExpAboveLevel:
+                $ref: '#/definitions/memberExpAboveLevel'
+              gameIds:
+                $ref: '#/definitions/gameIds'
```
ianbrown9475: These fields in the patch body should match the field names in the `member` resource

openapi.yaml
```diff
@@ -209,9 +438,324 @@ definitions:
...
+      - flop
+      - turn
+      - river
+      - showdown
+  selfLink:
+    properties:
+      self:
+        type: string
+        format: url
+        description: Self-link of current resource
```
ianbrown9475: Could you add a newline to the end of the file?

openapi.yaml
```diff
@@ -16,176 +16,405 @@ host: api.oregonstate.edu
...
+          schema:
+            $ref: '#/definitions/errors'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/errors'
+    delete:
+      tags:
+      - games
+      summary: "Remove a game specified by its id"
```
ianbrown9475: You should add something either here or in a description about how deleting the game will delete its players as well.

openapi.yaml
ianbrown9475: Should the `member` resource keep track of the ids of games that it is playing? This adds coupling between the `member` and the `game`. If you need to get all games that a member belongs to, you can do `GET /games?playerId={playerId}`.

ianbrown9475 (changes requested):
Nice work. As @tsoliangwu0130 mentioned, top-level objects in definitions should be upper camel case and properties of objects should be lower camel case.

openapi.yaml
nosnatef: The field would be fetched by the same possible query used in `GET /games?playerId={playerId}`

openapi.yaml
```diff
@@ -208,10 +435,332 @@ definitions:
...
+            $ref: '#/definitions/PlayerType'
+          attributes:
+            type: object
+            properties:
+              playerBet:
+                $ref: '#/definitions/PlayerBet'
+              playerStatus:
+                $ref: '#/definitions/PlayerStatus'
+              playerCards:
+                $ref: '#/definitions/Cards'
```
ianbrown9475, line 657: I think you want `additionalProperties: false` to be consistent with your other PATCH bodies

ianbrown9475 (changes requested):
Nice work on the changes. It looks like you made a commit on this branch that was changing the DAO and some other files. If you want to start making new changes on a different ticket, you should check out a new branch and make those changes there. This makes it easy to organize things and will save a lot of time if you need to go back and find where changes happened in previous PRs. Could you make sure these changes are committed on a different branch and then go back and revert them on this branch?

nosnatef:
I reverted the commit and fix the last problem.

ianbrown9475 (approved):

api/v1/serializers/pets-serializer.js
```diff
@@ -5,7 +5,7 @@ const _ = require('lodash');
 
 const { serializerOptions } = appRoot.require('utils/jsonapi');
 const { openapi } = appRoot.require('utils/load-openapi');
-const { paginate } = appRoot.require('utils/paginator');
+//const { paginate } = appRoot.require('utils/paginator');
```
ianbrown9475: Looks like this change was included by accident. This is causing a linter error:
```
   8:1   error  Expected exception block, space or tab after '//' in comment  spaced-comment
  43:22  error  'paginate' is not defined                                     no-undef
```

ianbrown9475 (changes requested):

openapi.yaml
```diff
@@ -208,10 +435,345 @@ definitions:
...
+          - type
+          - attributes
+        additionalProperties: false
+    required: 
+      - data
+    additionalProperties: false 
+  MemberResource:
+    properties:
+      id:
+        $ref: '#/definitions/PlayerId'
```
ianbrown9475, line 804: Since a `member` can be a part of multiple `games`, the `member` resource will have to have a different id than the `player` resource. @tsoliangwu0130 suggested the player resource should have an id like `{gameId}-{memberId}` or something similar.

ianbrown9475 (changes requested):

openapi.yaml
```diff
@@ -208,10 +441,348 @@ definitions:
...
+    type: string
+    example:  hunter2
+    format: password
+  MemberPatchBody:
+    properties:
+      data:
+        type: object
+        properties:
+          id:
+            $ref: '#/definitions/PlayerId'
```
ianbrown9475: I think this should be `MemberId`

openapi.yaml
```diff
@@ -208,10 +441,348 @@ definitions:
...
+          - type
+          - attributes
+        additionalProperties: false
+    required: 
+      - data
+    additionalProperties: false 
+  MemberResource:
+    properties:
+      id:
+        $ref: '#/definitions/PlayerId'
```
ianbrown9475, line 804: I think this should be `MemberId`

openapi.yaml
```diff
@@ -16,60 +16,343 @@ host: api.oregonstate.edu
...
+            $ref: '#/definitions/PlayerInGameResults'
+        '404':
+          description: Game not found
+          schema:
+            $ref: '#/definitions/Errors'
+        '500':
+          description: Internal server error
+          schema:
+            $ref: '#/definitions/Errors'
+  /games/{gameId}/players/{playerId}:
```
ianbrown9475, line 206: I think this should be `memberId`. There's a few other places that this will need to be changed also

openapi.yaml
```diff
@@ -208,10 +441,348 @@ definitions:
...
+    properties:
+      data:
+        type: object
+        properties:
+          type:
+            $ref: '#/definitions/GameType'
+          attributes:
+            type: object
+            properties:
+              playerIds:
```
ianbrown9475: I think when creating a new game, you'll want to specify a list of `memberIds` instead

openapi.yaml
```diff
@@ -208,10 +441,348 @@ definitions:
...
+          id:
+            $ref: '#/definitions/GameId'
+          type:
+            $ref: '#/definitions/GameType'
+          attributes:
+            type: object
+            properties:
+              round:
+                $ref: '#/definitions/Round'
+              playerIds:
```
ianbrown9475: I think when updating a game, you'll want to specify a list of `memberIds` instead

ianbrown9475 (changes requested):
Besides the places I commented on, there are a few other places where `playerid` will need to be changed to `memberId`

openapi.yaml
```diff
@@ -208,10 +497,378 @@ definitions:
...
+              - playerId
+            additionalProperties: false
+        required: 
+          - type
+          - attributes
+        additionalProperties: false
+    required:
+      - data
+    additionalProperties: false
+  PlayerResult:
```
ianbrown9475: It looks like you have `PlayerResult` and `PlayerInGameResult` that are duplicates of each other. You should be able to just use one of these. Since there is only one `player` resource, it might be better to have these called `PlayerResult` and `PlayerResource` instead of `PlayerInGameResult` and `PlayerInGameResource`

openapi.yaml
```diff
@@ -208,10 +497,378 @@ definitions:
...
+      data:
+        type: object
+        properties:
+          type:
+            $ref: '#/definitions/PlayerType'
+          attributes:
+            type: object
+            properties:
+              memberId:
+                $ref: '#/definitions/MemberId'
```
ianbrown9475, line 714: Should the post body also allow `playerBet`, `playerStatus`, and `playerCards`? You could make these optional and add default values for them if you want. This could make it easier to create a player with specific starting attributes.

openapi.yaml
```diff
@@ -208,10 +497,378 @@ definitions:
...
+            type: object
+            properties:
+              memberIds:
+                $ref: '#/definitions/MemberIds'
+              minimumBet:
+                $ref: '#/definitions/MinimumBet'
+              maximumBet:
+                $ref: '#/definitions/MaximumBet'
+            required:
+              - playerIds
```
ianbrown9475: I think this should be `memberIds`

openapi.yaml
```diff
@@ -208,10 +497,378 @@ definitions:
...
+        $ref: '#/definitions/GameId'
+      type:
+        $ref: '#/definitions/GameType'
+      attributes:
+        type: object
+        properties:
+          round:
+            $ref: '#/definitions/Round'
+          playerIds:
+            $ref: '#/definitions/PlayerIds'
```
ianbrown9475: I think this should be memberIds

openapi.yaml
```diff
@@ -208,10 +497,378 @@ definitions:
...
+    type: string
+    example: '101'
+  MemberIds:
+    type: array
+    items:
+      $ref: '#/definitions/MemberId'
+  PlayerId:
+    type: string
+    example: 'g1m1'
+  PlayerIds:
```
ianbrown9475: I think after you change playerIds to memberIds in the places I commented on, this definition won't be used anywhere and you should be able to delete it

ianbrown9475 (changes requested):

openapi.yaml
nosnatef: I didn't. The resource are requested and returned with playerId mostly on this endpoint.

openapi.yaml
nosnatef: That is what I decided to return after yesterday's discussion.

openapi.yaml
```diff
@@ -208,10 +497,372 @@ definitions:
...
+            type: object
+            properties:
+              memberIds:
+                $ref: '#/definitions/MemberIds'
+              minimumBet:
+                $ref: '#/definitions/MinimumBet'
+              maximumBet:
+                $ref: '#/definitions/MaximumBet'
+            required:
+              - playerIds
```
ianbrown9475: Looks like this should be `memberIds`

openapi.yaml
```diff
@@ -16,60 +16,399 @@ host: api.oregonstate.edu
...
+          name: round
+          type: string
+          description: The current round of the game
         - in: query
-          name: species
-          description: Search by pet species
-          required: false
+          name: memberId
           type: string
+          description: The memberId of members that are playing this game
```
ianbrown9475: Since the `game` resource doesn't have a `memberIds` property, could you remove this query parameter?

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #3, opened 2019-06-20
Feature/co1443 db design

nosnatef:


api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (43, 4, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (44, 5, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (45, 6, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (46, 7, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (47, 8, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (48, 9, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (49, 10, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (50, 11, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (51, 12, 4);
+INSERT INTO CARDS (CARD_ID, CARD_NUMBER_ID, SUIT_ID) VALUES (52, 13, 4);
```
ianbrown9475: You may want to combine all of these `INSERT` statements into a single `INSERT ALL` statement https://www.techonthenet.com/oracle/questions/insert_rows.php

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+
+CREATE TABLE CARD_NUMBERS (
+    CARD_NUMBER_ID NUMBER NOT NULL,
+    CARD_NUMBER    VARCHAR2(255 BYTE),
+    PRIMARY KEY (CARD_NUMBER_ID)
+);
+
+COMMENT ON TABLE CARD_NUMBERS IS 'Contains possible values for card number';
+COMMENT ON COLUMN CARD_NUMBERS.CARD_NUMBER_ID IS 'ID of number';
+COMMENT ON COLUMN CARD_NUMBERS.CARD_NUMBER IS 'number name';
```
ianbrown9475, line 36: Nice job adding comments, could you also do this for the other columns in the other tables?

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+);
+
+COMMENT ON TABLE CARD_SUITS IS 'Contains possible values for card suit';
+COMMENT ON COLUMN CARD_SUITS.SUIT_ID IS 'ID of suit';
+COMMENT ON COLUMN CARD_SUITS.SUIT IS 'Suit name';
+
+INSERT INTO CARD_SUITS (SUIT_ID, SUIT) VALUES (1,'spades');
+INSERT INTO CARD_SUITS (SUIT_ID, SUIT) VALUES (2,'diamonds');
+INSERT INTO CARD_SUITS (SUIT_ID, SUIT) VALUES (3,'clubs');
+INSERT INTO CARD_SUITS (SUIT_ID, SUIT) VALUES (4,'hearts');
```
ianbrown9475: @tsoliangwu0130 mentioned that Banner tables will use IDs like `'E'` for employed and `'T'` for terminated and have another column with a description field. Using something like this would be more meaningful than `1`, `2`, `3`, and `4`

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
+DROP TABLE CARD_SUITS;
+DROP TABLE CARD_NUMBERS;
+DROP TABLE GAMES;
+DROP TABLE MEMBERS;
+DROP TABLE PARTICIPATIONS;
+DROP TABLE ROUNDS;
+DROP TABLE STATUSES;
```
ianbrown9475: These statements don't drop all of the tables created in the script. Also, the order needs to be done in a way that will drop all tables with foreign key constraints before the tables being referenced. An easy way to do this is to drop all of the tables in the reverse order that they were declared.

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+    BET_POOL NUMBER NOT NULL,
+    PRIMARY KEY (GAME_ID),
+    FOREIGN KEY (ROUND_ID) REFERENCES ROUNDS(ROUND_ID)
+);
+
+CREATE TABLE TABLE_CARDS (
+    OWNERSHIP_ID NUMBER NOT NULL,
+    GAME_ID NUMBER NOT NULL,
+    CARD_ID NUMBER NOT NULL,
+    PRIMARY KEY (OWNERSHIP_ID),
```
ianbrown9475: For this table and other simple relationship tables that consist of two IDs, you shouldn't need to create a third ID to uniquely identify the relationship. Instead, you could have the primary key be a composite key of the two ids. For example here, you could have `GAME_ID` and `CARD_ID` together be a composite key for the table. This also prevents duplicate relationships.

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+
+CREATE TABLE PARTICIPATIONS (
+    PARTICIPATION_ID NUMBER NOT NULL,
+    MEMBER_ID NUMBER NOT NULL,
+    GAME_ID NUMBER NOT NULL,
+    PRIMARY KEY (PARTICIPATION_ID),
+    FOREIGN KEY (MEMBER_ID) REFERENCES MEMBERS(MEMBER_ID),
+    FOREIGN KEY (GAME_ID) REFERENCES GAMES(GAME_ID)
+);
+
```
ianbrown9475, line 232: It looks like you've made all of your `NUMBER` columns not nullable but all of your `VARCHAR2` columns nullable. Is there a reason for this? Just curious.

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,192 @@
...
+CREATE TABLE PARTICIPATIONS (
+    PARTICIPATION_ID NUMBER NOT NULL,
+    MEMBER_ID NUMBER NOT NULL,
+    GAME_ID NUMBER NOT NULL,
+    PRIMARY KEY (PARTICIPATION_ID),
+    FOREIGN KEY (MEMBER_ID) REFERENCES MEMBERS(MEMBER_ID),
+    FOREIGN KEY (GAME_ID) REFERENCES GAMES(GAME_ID)
+);
```
ianbrown9475: Is it necessary to have this table for keeping track of participations? I think you should be able to have a column on `PLAYERS` that references the game that they belong to since a player shouldn't be able to exist without a game

ianbrown9475 (changes requested):

api/v1/db/oracledb/LearningDB.sql
nosnatef: It's 52 lines of code I generated using Python and insert all does not greatly reduce the code amount so I don't think I need to change it

ianbrown9475: The main benefit of using an `INSERT ALL` statement would be to use 1 database query instead of 52 database queries to insert the rows. This wouldn't reduce the code size, but would reduce insertion time and database traffic. This isn't necessary though, just something to keep in mind.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "ENV/bin/python3.7"
+}
```
ianbrown9475: Could you remove this file from git? If you want to avoid committing this directory, you can add it to `.gitignore`

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,232 @@
...
+    UNIQUE (MEMBER_NICKNAME)
+);
+
+COMMENT ON TABLE MEMBERS IS 'The member in the system';
+COMMENT ON COLUMN MEMBERS.MEMBER_ID IS 'The ID of the member';
+COMMENT ON COLUMN MEMBERS.MEMBER_NICKNAME IS 'The user name of the member';
+COMMENT ON COLUMN MEMBERS.MEMBER_EMAIL IS 'The email address of the member';
+COMMENT ON COLUMN MEMBERS.MEMBER_LEVEL IS 'The level of member in the system';
+COMMENT ON COLUMN MEMBERS.MEMBER_EXP_OVER_LEVEL IS 'The experience of member that is above the current level';
+COMMENT ON COLUMN MEMBERS.PASSWORD IS 'The password of member, which should be stored in hashed form';
```
ianbrown9475: I think this should be `MEMBERS.MEMBER_PASSWORD`

ianbrown9475 (changes requested):

api/v1/db/oracledb/LearningDB.sql
```diff
@@ -0,0 +1,232 @@
...
+
+INSERT INTO MEMBERS (MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL, MEMBER_PASSWORD) VALUES (1, 'guz', 'guz@oregonstate.edu', 123, 10492, 'hunter3');
+INSERT INTO MEMBERS (MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL, MEMBER_PASSWORD) VALUES (2, 'John Wick', 'wickj@oregonstate.edu', 200, 114514, 'hunter10');
```
ianbrown9475, line 156: Also since this looks like it's just test data, could you add it to the bottom as a SQL comment or remove it from the script?

nosnatef: I will remove it after the endpoints are implemented.

ianbrown9475 (approved):

PR #4, opened 2019-06-25
Feature/co1469 get member

nosnatef:


api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * @summary Return a list of members
+ * @function
+ * @returns {Promise<Object[]>} Promise object represents a list of members
+ */
+const getMembers = async (query) => {
+  const connection = await conn.getConnection();
+  try {
+    const rawMembers = await connection.execute('SELECT * FROM MEMBERS');
```
ianbrown9475: Even though you may want all fields from the `MEMBERS` table, it's good practice to specify all of the columns in the `SELECT` statement instead of `SELECT *`. Here's an example: https://github.sig.oregonstate.edu/ecs-data/persons-api-contrib/blob/master/AbstractPersonsDAO.groovy

ianbrown9475: I think this endpoint has query parameters used to filter the results. Could you implement these?

ianbrown9475: The return of `connection.execute()` will be an object similar to this
```json
{
  "metadata": {...},
  "rows": [...]
}
```
Could you rename `rawMembers` to fit this?
You could also do something like
```javascript
const { rows: rawMembers } = await connection.execute(...);
```

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,86 @@
...
+   */
+
+  // const pageQuery = {
+  //   size: query['page[size]'],
+  //   number: query['page[number]'],
+  // };
+
+  // const pagination = paginate(rawMembers, pageQuery);
+  // pagination.totalResults = rawMembers.length;
+  // rawMembers = pagination.paginatedRows;
```
ianbrown9475: I think you can delete this pagination code since `GET /members` won't be paginated

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const topLevelSelfLink = paramsLink(memberResourceUrl, query);
+  const serializerArgs = {
+    identifierField: 'MEMBER_ID',
+    resourceKeys: memberResourceKeys,
+    resourcePath: memberResourcePath,
+    topLevelSelfLink,
+    // query: _.omit(query, 'page[size]', 'page[number]'),
+    query,
+    enableDataLinks: true,
```
ianbrown9475, line 51: I mentioned this before, but there is a parameter called `keyForAttribute` that can be used which will handle converting field names in the database to the field names referenced by the openapi resource. https://github.com/SeyZ/jsonapi-serializer#available-serialization-option-opts-argument. Since the names in the database don't exactly line up with the names in the specification, you will need to write a function to handle this.

This function should basically do the reverse of what the code here https://github.com/osu-mist/poker-api/blob/8b40a2fdc8e5bcffaac0f643bf90def144f109a4/api/v1/serializers/members-serializer.js#L17-L25 ends up doing.

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,86 @@
...
+  // rawMembers = pagination.paginatedRows;
+
+  const topLevelSelfLink = paramsLink(memberResourceUrl, query);
+  const serializerArgs = {
+    identifierField: 'MEMBER_ID',
+    resourceKeys: memberResourceKeys,
+    resourcePath: memberResourcePath,
+    topLevelSelfLink,
+    // query: _.omit(query, 'page[size]', 'page[number]'),
+    query,
```
ianbrown9475, line 50: This `query` parameter is only used for pagination purposes so you should be able to delete it

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * The column name getting from database is usually UPPER_CASE.
+ * This block of code is to make the camelCase keys defined in openapi.yaml be
+ * UPPER_CASE so that the serializer can correctly match the corresponding columns
+ * from the raw data rows.
+ */
+_.forEach(memberResourceKeys, (key, index) => {
+  memberResourceKeys[index] = decamelize(key).toUpperCase();
+});
```
ianbrown9475, line 25: This function simply converts `lowerCamelCase` keys into `UPPER_SNAKE_CASE` keys. Since your openapi names and database column names don't use exactly the same names, you will need to modify this function so that the keys are correctly converted to your database column names

ianbrown9475 (changes requested):
Here's an example response I got:
```json
{
    "links": {
        "self": "https://localhost/v1/members"
    },
    "data": [
        {
            "type": "member",
            "id": "1",
            "links": {
                "self": "https://localhost/v1/members/1"
            },
            "attributes": {
                "memberEmail": "test@test.com",
                "memberLevel": "1",
                "memberExpOverLevel": "0"
            }
        }
    ]
}
```
It looks like some fields are missing from the response. I made some comments on the member serializer that should address this. Also, `memberLevel` and `memberExpOverLevel` should be integers as defined in your openapi document but are being returned as strings.

nosnatef:
Still need to make sure some fields returned to be integer form. Also, working on handling the query params.

nosnatef:
Solved the type issue at the local repo.

api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+    if (memberEmail) {
+      sqlParams.memberEmail = memberEmail;
+    }
+    const sqlQuery = `
+    SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL FROM MEMBERS 
+    WHERE 1 = 1 
+    ${memberName ? 'AND MEMBER_NICKNAME = :memberName ' : ''} 
+    ${memberEmail ? 'AND MEMBER_EMAIL = :memberEmail' : ''}
+    `;
```
ianbrown9475, line 31: Could you update this query to get the game ids for the member since `gameIds` should be a property on the response?

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #5, opened 2019-06-26
Feature/co1470 get member by id

nosnatef:


api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+const getMemberById = async (id) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlQuery = `
+    SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL 
+    FROM MEMBERS WHERE MEMBER_ID = :id
+    `;
+    const sqlParams = [id];
+    console.log(sqlQuery, sqlParams);
```
ianbrown9475: Could you remove this `console.log()` statement?

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,77 @@
...
+  ).serialize(rawMembers);
+};
+
+/**
+ * @summary Serialize memberResource to JSON API
+ * @function
+ * @param {Object} rawMember Raw data row from data source
+ * @returns {Object} Serialized memberResource object
+ */
+const serializeMember = (rawMember) => {
```
ianbrown9475, line 65: You'll need to convert the types to match the openapi document like you do in `GET /members`. Since that conversion logic will be shared by both `serializeMembers` and `serializeMember`, can you put it into a function?

api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {Promise<Object>} Promise object represents a specific member
+ *                            or return undefined if term
+ *                            is not found
+ */
+const getMemberById = async (id) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlQuery = `
+    SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL 
+    FROM MEMBERS WHERE MEMBER_ID = :id
```
ianbrown9475, line 55: I tried making this request `GET /members/a` and got a `500` response back. This was printed to the console:
```
SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL
    FROM MEMBERS WHERE MEMBER_ID = :id
     [ 'a' ]
Error: ORA-01722: invalid number
```
You define `memberId` to be a string in your openapi document so non-number strings are allowed. I think you want to change the type of this field in your database script to be a `VARCHAR2`

ianbrown9475 (changes requested):

api/v1/db/oracledb/member-dao.js
nosnatef: I will change it to be a number.

nosnatef:
Just wondering why does the topLink from /member/:id look like this?
`{
    "links": {
        "self": "https://localhost/v1/members/undefined"
    },
    "data": [
        {
            "type": "member",
            "id": "1",
            "links": {
                "self": "https://localhost/v1/members/1"
            },
            "attributes": {
                "memberNickname": "guz",
                "memberEmail": "guz@oregonstate.edu",
                "memberLevel": 123,
                "memberExpOverLevel": 10492
            }
        }
    ]
}`

nosnatef:
The self link should not be ..../undefined, right?

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,90 @@
...
+ * @summary Serialize memberResources to JSON API
+ * @function
+ * @param {[Object]} rawMembers Raw data rows from data source
+ * @param {Object} query Query parameters
+ * @returns {Object} Serialized memberResources object
+ */
+const serializeMembers = (rawMembers, query) => {
+  /**
+   * Add pagination links and meta information to options if pagination is enabled
+   */
```
ianbrown9475: Can you remove this comment since the code that it was referencing has been deleted?

api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,74 @@
...
+    const sqlParams = [id];
+    const rawMembersResponse = await connection.execute(sqlQuery, sqlParams);
+    const rawMembers = rawMembersResponse.rows;
+    if (_.isEmpty(rawMembers)) {
+      return undefined;
+    }
+    if (rawMembers.length > 1) {
+      throw new Error('Expect a single object but got multiple results.');
+    } else {
+      const serializedMember = serializeMember(rawMembers);
```
ianbrown9475: `serializeMember` expects a single object `rawMember` but you are passing an array. I think fixing this will also fix the top level self link issue you mentioned here https://github.com/osu-mist/poker-api/pull/5#issuecomment-507797862

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+/**
+ * @summary Serialize memberResource to JSON API
+ * @function
+ * @param {Object} rawMember Raw data row from data source
+ * @returns {Object} Serialized memberResource object
+ */
+const serializeMember = (rawMember) => {
+  memberConverter(rawMember);
```
ianbrown9475: `memberConverter` is expecting an array of objects but you will be passing it a single object once you fix the dao. You will need to fix this after the other change

openapi.yaml
```diff
@@ -427,9 +450,9 @@ parameters:
   memberId:
     name: memberId
     in: path
-    description: unique string
+    description: unique integer
     required: true
-    type: string
+    type: integer
```
ianbrown9475, line 455: You will also need to update `Definitions.MemberId` now that it is a number instead of a string

ianbrown9475 (changes requested):

api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,74 @@
...
+const getMemberById = async (id) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlQuery = `
+    SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL 
+    FROM MEMBERS WHERE MEMBER_ID = :id
+    `;
+    const sqlParams = [id];
+    const rawMembersResponse = await connection.execute(sqlQuery, sqlParams);
+    const [rawMembers] = rawMembersResponse.rows;
```
ianbrown9475: The value of `rawMembers` is being set to the first item in the array and if the array is empty, it will become `undefined`. Since there's code after this that checks whether `rawMembers` is empty and if its length is more than 1, `rawMembers` should be the array of sql rows returned from the query. After the checks on the rows have been done, you can call the serializer on the first item in the list and return the result

api/v1/db/oracledb/member-dao.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {string} id Unique member ID
+ * @returns {Promise<Object>} Promise object represents a specific member
+ *                            or return undefined if term
+ *                            is not found
+ */
+const getMemberById = async (id) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlQuery = `
+    SELECT MEMBER_ID, MEMBER_NICKNAME, MEMBER_EMAIL, MEMBER_LEVEL, MEMBER_EXP_OVER_LEVEL
```
ianbrown9475: It looks like there is trailing whitespace on this line. Could you remove it? If you're using vscode, you can add this to your `settings.json`:
```json
"files.trimFinalNewlines": true,
"files.trimTrailingWhitespace": true
```
and it should take care of this for you whenever you save files

api/v1/serializers/members-serializer.js
```diff
@@ -0,0 +1,89 @@
...
+  _.forEach(rawMembers, (member) => {
+    member.MEMBER_LEVEL = parseInt(member.MEMBER_LEVEL, 10);
+    member.MEMBER_EXP_OVER_LEVEL = parseInt(member.MEMBER_EXP_OVER_LEVEL, 10);
+  });
+};
+
+const singleMemberConverter = (rawMember) => {
+  rawMember.MEMBER_LEVEL = parseInt(rawMember.MEMBER_LEVEL, 10);
+  rawMember.MEMBER_EXP_OVER_LEVEL = parseInt(rawMember.MEMBER_EXP_OVER_LEVEL, 10);
+};
```
ianbrown9475, line 29: It looks like the logic for `memberConverter` and `singleMemberConverter` is almost identical. You should be able to combine this into a single function. You might have to change how you call it though

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #8, opened 2019-06-27
Add new endpoint

nosnatef:


openapi.yaml
```diff
@@ -334,6 +334,29 @@ paths:
           description: Internal server error
           schema:
             $ref: '#/definitions/Errors'
+  /members/{memberId}/games:
+    get:
+      summary: Info for games that the member is current playing
+      tags:
+        - members
```
ianbrown9475: I think this should be tagged as games since the response will be games, right?

openapi.yaml
```diff
@@ -334,6 +334,29 @@ paths:
           description: Internal server error
           schema:
             $ref: '#/definitions/Errors'
+  /members/{memberId}/games:
```
ianbrown9475: Could you put this after `/members/{memberId}`?

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #9, opened 2019-06-28
Feature/co1480 get games

nosnatef:


api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
...
+      mergedRawGames[mergedRawGames.length - 1].tableCards.push({
+        cardNumber: rawGames[i].CARD_NUMBER,
+        cardSuit: rawGames[i].SUIT,
+      });
+    }
+  }
+  return mergedRawGames;
+};
+
+const getGames = async (query) => {
```
ianbrown9475, line 13: Since you're exporting this function to be used by other modules, could you add documentation similar to here? https://github.com/osu-mist/poker-api/blob/4dda0826776cd3728e3e08ab6f8aaa0f21e03ef7/api/v1/db/oracledb/member-dao.js#L9-L14
This documentation is called JSDoc https://jsdoc.app/

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
+const appRoot = require('app-root-path');
+
+const { serializeGames } = require('../../serializers/games-serializer');
+
+const conn = appRoot.require('api/v1/db/oracledb/connection');
```
ianbrown9475, line 5: Could you add a blank line separating this variable from the function declaration?

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
...
+const mergeRawGames = (rawGames) => {
+  const mergedRawGames = [];
+  let currentId = 'dummy';
```
ianbrown9475: Instead of setting this to an actual value, it would be better practice to set it to something like `undefined` or `null` and then test that `currentId !== undefined` or `null` explicitly.

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
...
+    SELECT CARD_NUMBERS.CARD_NUMBER, CARD_SUITS.SUIT, GAMES.GAME_ID, ROUNDS.ROUND, GAMES.MAXIMUM_BET, 
+    GAMES.MINIMUM_BET, GAMES.BET_POOL 
+    FROM TABLE_CARDS, GAMES, CARDS, CARD_SUITS, CARD_NUMBERS, ROUNDS 
+    WHERE GAMES.GAME_ID = TABLE_CARDS.GAME_ID AND 
+    TABLE_CARDS.GAME_ID = GAMES.GAME_ID AND 
+    TABLE_CARDS.CARD_ID = CARDS.CARD_ID AND
+    CARDS.CARD_NUMBER_ID = CARD_NUMBERS.CARD_NUMBER_ID AND 
+    CARDS.CARD_SUIT_ID = CARD_SUITS.SUIT_ID AND 
+    GAMES.ROUND_ID = ROUNDS.ROUND_ID
+    ${roundId ? 'AND ROUND_ID = :roundId' : ''}
```
ianbrown9475: Right now, this is not doing anything since the query parameter for `GET /games` is called `round` and should specify the name of the round, not the round ID

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
...
+    const sqlQuery = `
+    SELECT CARD_NUMBERS.CARD_NUMBER, CARD_SUITS.SUIT, GAMES.GAME_ID, ROUNDS.ROUND, GAMES.MAXIMUM_BET, 
+    GAMES.MINIMUM_BET, GAMES.BET_POOL 
+    FROM TABLE_CARDS, GAMES, CARDS, CARD_SUITS, CARD_NUMBERS, ROUNDS 
+    WHERE GAMES.GAME_ID = TABLE_CARDS.GAME_ID AND 
+    TABLE_CARDS.GAME_ID = GAMES.GAME_ID AND 
+    TABLE_CARDS.CARD_ID = CARDS.CARD_ID AND
+    CARDS.CARD_NUMBER_ID = CARD_NUMBERS.CARD_NUMBER_ID AND 
+    CARDS.CARD_SUIT_ID = CARD_SUITS.SUIT_ID AND 
+    GAMES.ROUND_ID = ROUNDS.ROUND_ID
```
ianbrown9475: It looks like this main query is getting the needed information, but the query itself is difficult to read. It's usually best practice to select the main table, in this case `GAMES`, and join the additional tables using some form of `JOIN` such as `INNER JOIN`, `RIGHT JOIN`, etc. The query should also use indentation and line breaks to improve readability.

Also, you can use table aliases if you want to shorten some of the code. Example:
```sql
SELECT L.PROP1, L.PROP2, L.PROP3
FROM LONG_TABLE_NAME L
```

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,58 @@
...
+  for (let i = 0; i < rawGames.length; i += 1) {
+    if (rawGames[i].GAME_ID !== currentId) {
+      currentId = rawGames[i].GAME_ID;
+      rawGames[i].tableCards = [{
+        cardNumber: rawGames[i].CARD_NUMBER,
+        cardSuit: rawGames[i].SUIT,
+      }];
+      mergedRawGames.push(rawGames[i]);
+    } else {
+      mergedRawGames[mergedRawGames.length - 1].tableCards.push({
```
ianbrown9475: This function relies on the rows being sorted by their `GAME_ID`. You should be able to write this logic in a way that doesn't need this restriction, in case your query or database configuration changes in the future.

I would look into the Lodash package. This is included in our express api dependencies by default and contains many helpful functional programming methods that can simplify your code. One method that might be helpful here is `_.groupBy()` https://lodash.com/docs/4.17.11#groupBy.

ianbrown9475 (changes requested):

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
+const appRoot = require('app-root-path');
+const _ = require('lodash');
+
+
```
ianbrown9475, line 2: You can remove this extra line, in my previous comment, I meant add a blank line separating `const conn = ...` and `const mergeRawGames = ...`

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+const { serializeGames } = require('../../serializers/games-serializer');
+
+const conn = appRoot.require('api/v1/db/oracledb/connection');
+const mergeRawGames = (rawGames) => {
+  const groupedRawGames = _.groupBy(rawGames, 'GAME_ID');
+  const mergedRawGames = _.map(groupedRawGames, (gameMetaDataArray) => {
+    if (gameMetaDataArray[0].CARD_NUMBER == null) {
+      gameMetaDataArray[0].tableCards = [];
+    } else {
+      gameMetaDataArray[0].tableCards = _.map(gameMetaDataArray, (data) => {
```
ianbrown9475: You could use the ternary operator here instead of the `if..else` statement:
```js
gameMetaDataArray[0].tableCards = gameMetaDataArray[0].CARD_NUMBER === null ? [] : _.map(...)
```

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+const _ = require('lodash');
+
+
+const { serializeGames } = require('../../serializers/games-serializer');
+
+const conn = appRoot.require('api/v1/db/oracledb/connection');
+const mergeRawGames = (rawGames) => {
+  const groupedRawGames = _.groupBy(rawGames, 'GAME_ID');
+  const mergedRawGames = _.map(groupedRawGames, (gameMetaDataArray) => {
+    if (gameMetaDataArray[0].CARD_NUMBER == null) {
```
ianbrown9475: Could you use `===` here instead of `==`? This will make sure the actual value is `null` instead of something like `undefined` for example
```js
undefined == null
// true
undefined === null
// false
```

api/v1/db/oracledb/game-dao.js
ianbrown9475: I got a linter error here
```
14:76  error  Unexpected block statement surrounding arrow body; parenthesize the returned value and move it immediately after the `=>`  arrow-body-style
```
Remember to commit without disabling pre-commit scripts unless you need to make a work-in-progress change that won't be up for a pull request yet

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+      gameMetaDataArray[0].tableCards = [];
+    } else {
+      gameMetaDataArray[0].tableCards = _.map(gameMetaDataArray, (data) => {
+        return {
+          cardNumber: data.CARD_NUMBER,
+          cardSuit: data.SUIT,
+        };
+      });
+    }
+    return (gameMetaDataArray[0]);
```
ianbrown9475: I think these parentheses are unnecessary

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+  });
+  return mergedRawGames;
+};
```
ianbrown9475: Nice work on this function. Since this logic is pretty complex, could you add some documentation on what it is doing?

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+    const sqlQuery = `
+    SELECT CN.CARD_NUMBER, CS.SUIT, G.GAME_ID, R.ROUND, G.MAXIMUM_BET, 
+    G.MINIMUM_BET, G.BET_POOL 
+    FROM GAMES G
+    INNER JOIN ROUNDS R ON G.ROUND_ID = R.ROUND_ID 
+    LEFT OUTER JOIN TABLE_CARDS TC ON TC.GAME_ID = G.GAME_ID 
+    LEFT OUTER JOIN CARDS C ON TC.CARD_ID = C.CARD_ID 
+    LEFT OUTER JOIN CARD_NUMBERS CN ON C.CARD_NUMBER_ID = CN.CARD_NUMBER_ID 
+    LEFT OUTER JOIN CARD_SUITS CS ON C.CARD_SUIT_ID = CS.SUIT_ID 
+    ${round ? 'AND ROUNDS.ROUND = :round' : ''}
```
ianbrown9475: I tried `GET /games?round=blind` and I got this error:
```
Error: ORA-00904: "ROUNDS"."ROUND": invalid identifier
```
Could you look into this?

ianbrown9475 (changes requested):

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,63 @@
+const appRoot = require('app-root-path');
+const _ = require('lodash');
+
+const { serializeGames } = require('../../serializers/games-serializer');
+
+const conn = appRoot.require('api/v1/db/oracledb/connection');
+
+/**
+ * @summary Merge the raw response from database. Extract card information from every row,
+ * and put them into a field called 'tableCard' in the individual merged game object, while other
```
ianbrown9475: Typo 'tableCard' -> 'tableCards'

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,63 @@
...
+    return gameMetaDataArray[0];
+  });
+  return mergedRawGames;
+};
+
+/**
+ * @summary Return a list of games
+ * @function
+ * @returns {Promise<Object[]>} Promise object represents a list of games
+ */
```
ianbrown9475, line 12: Could you add `@param` to this function and `mergeRawGames`? I think in VS code, you can add some starter documentation by placing your cursor on the line above the function definition and typing `/**[enter]` which can be helpful

ianbrown9475 (changes requested):

api/v1/db/oracledb/game-dao.js
```diff
@@ -6,11 +6,9 @@ const { serializeGames } = require('../../serializers/games-serializer');
...
 
 /**
- * @summary Merge the raw response from database. Extract card information from every row,
- * and put them into a field called 'tableCard' in the individual merged game object, while other
- * properties in the object remained in the first layer.
- * @function
- * @returns {Promise<Object[]>} Promise object represents a list of games
+ * @summary Return a list of games merged by GAME_ID property.
+ * @param {[Object]} rawGames An array of raw game data returned from SQL database.
+ * @returns {[Object]} Game objects merged.
```
ianbrown9475: Could you add back your longer description in the summary field, add back `@function` and fix the types on `rawGames` and `returns`?

api/v1/db/oracledb/game-dao.js
```diff
@@ -27,7 +25,7 @@ const mergeRawGames = (rawGames) => {
 
 /**
  * @summary Return a list of games
- * @function
```
ianbrown9475: Could you add back `@function`?

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,61 @@
...
+  const mergedRawGames = _.map(groupedRawGames, (gameMetaDataArray) => {
+    gameMetaDataArray[0].tableCards = gameMetaDataArray[0].CARD_NUMBER === null ? []
+      : _.map(gameMetaDataArray, data => ({
+        cardNumber: data.CARD_NUMBER,
+        cardSuit: data.SUIT,
+      }));
+    return gameMetaDataArray[0];
+  });
+  return mergedRawGames;
+};
```
ianbrown9475: Now that I think about it, `mergeRawGames` should belong in the games serializer. The purpose of the DAO should be just to take information about a query and fetch data from the data source without any serialization or modification. The serializer should handle converting the raw data to the serialized response data. Doing some of the serialization here and the rest in the serializer doesn't seem that intuitive.

ianbrown9475 (changes requested):
I noticed that the value of `round` on the games resource is an enum of a few possible values, but the query parameter to filter games by can be any string. Does this make sense? It might be better to either make the query parameter an enum with the same values or make the round property a freeform string. What do you think?

api/v1/serializers/games-serializer.js
```diff
@@ -0,0 +1,75 @@
...
+  });
+};
+
+/**
+ * @summary Merge the raw response from database. Extract card information from every row,
+ * and put them into a field called 'tableCards' in the individual merged game object, while other
+ * properties in the object remained in the first layer.
+ * @function
+ * @param {Array[Object]} rawGames An array of raw game data returned from SQL database.
+ * @returns {Array[Object]} Game objects merged.
```
ianbrown9475: I think these types should be `{Object[]}`

openapi.yaml
```diff
@@ -57,11 +57,17 @@ paths:
...
         - $ref: '#/parameters/authorization'
         - in: query
           name: round
           type: string
+          enum:
+            - blind
+            - flop
+            - turn
+            - river
+            - showdown
```
ianbrown9475, line 70: Could you use a YAML anchor for the enum like we discussed so you don't need to repeat the code here? https://swagger.io/docs/specification/2-0/enums/

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #10, opened 2019-07-02
Feature/co1481 get game by id

nosnatef:

PR #11, opened 2019-07-02
Feature/co1482 get games by memberid

nosnatef:


api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,121 @@
...
+const { serializeGames, serializeGame } = require('../../serializers/games-serializer');
+
+const conn = appRoot.require('api/v1/db/oracledb/connection');
+const mergeRawGames = (rawGames) => {
+  const groupedRawGames = _.groupBy(rawGames, 'GAME_ID');
+  const mergedRawGames = _.map(groupedRawGames, (gameMetaDataArray) => {
+    if (gameMetaDataArray[0].CARD_NUMBER == null) {
+      gameMetaDataArray[0].tableCards = [];
+    } else {
+      gameMetaDataArray[0].tableCards = _.map(gameMetaDataArray, (data) => {
```
ianbrown9475: I got a linting error here: `13:76  error  Unexpected block statement surrounding arrow body; parenthesize the returned value and move it immediately after the `=>`  arrow-body-style`

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,121 @@
...
+    GAMES.MINIMUM_BET, GAMES.BET_POOL 
+    FROM TABLE_CARDS, GAMES, CARDS, CARD_SUITS, CARD_NUMBERS, ROUNDS, PLAYERS
+    WHERE GAMES.GAME_ID = TABLE_CARDS.GAME_ID AND 
+    TABLE_CARDS.GAME_ID = GAMES.GAME_ID AND 
+    TABLE_CARDS.CARD_ID = CARDS.CARD_ID AND
+    CARDS.CARD_NUMBER_ID = CARD_NUMBERS.CARD_NUMBER_ID AND 
+    CARDS.CARD_SUIT_ID = CARD_SUITS.SUIT_ID AND 
+    GAMES.ROUND_ID = ROUNDS.ROUND_ID AND 
+    PLAYERS.MEMBER_ID = :id AND 
+    PLAYERS.GAME_ID = GAMES.GAME_ID
```
ianbrown9475: From our discussion, could you update this query to use joins?
Also, since it should end up looking similar to the other two sql queries for games, could you refactor?

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,121 @@
...
+    `;
+    const rawGamesResponse = await connection.execute(sqlQuery, sqlParams);
+    let rawGames = rawGamesResponse.rows;
+    rawGames = mergeRawGames(rawGames);
+    console.log(rawGames);
```
ianbrown9475: Could you remove this `console` statement?

ianbrown9475 (changes requested):
* I queried `GET /members/999/games` where 999 is a member id that doesn't exist and the response was
  ```json
  {
      "links": {
          "self": "https://localhost/v1/games"
      },
      "data": []
  }
  ```
  but the expected response should be a 404
* The top-level self link seems to be `https://localhost/v1/games` which is incorrect
* I queried `GET /members/a/games` and got a 500 with this printed to the console: `Error: ORA-01722: invalid number`

Could you look into these issues?

api/v1/serializers/games-serializer.js
```diff
@@ -0,0 +1,76 @@
...
+  rawGame.BET_POOL = parseInt(rawGame.BET_POOL, 10);
+};
+
+const serializeGames = (rawGames, query, memberId) => {
+  _.forEach(rawGames, (game) => {
+    individualGameConverter(game);
+  });
+
+  const topLevelSelfLink = !memberId ? paramsLink(gameResourceUrl, query)
+    : `${resourcePathLink(memberResourceUrl, memberId)}/games`;
```
ianbrown9475: Could you wrap this second string in `paramsLink()`? This will add any query parameters to the self link.

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,107 @@
...
+    connection.close();
+  }
+};
+
+const getGamesByMemberId = async (id, query) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlParams = [id];
+    const getMemberSqlQuery = `SELECT * FROM MEMBERS M
+    WHERE M.MEMBER_ID = :id
```
ianbrown9475, line 48: From our discussion, since we don't need the values of the response here, could you use something like `SELECT COUNT(1) FROM MEMBERS M WHERE M.MEMBER_ID = :id`? Then instead of checking that the rows are not empty, you would want to check that the value of `COUNT(1)` is not 0.

api/v1/db/oracledb/game-dao.js
```diff
@@ -0,0 +1,107 @@
...
+    if (_.isEmpty(rawMembers)) {
+      return undefined;
+    }
+    const getSqlQuery = `${sqlQuery}
+    LEFT OUTER JOIN PLAYERS P ON P.GAME_ID = G.GAME_ID
+    WHERE P.MEMBER_ID = :id
+    `;
+    const rawGamesResponse = await connection.execute(getSqlQuery, sqlParams);
+    let rawGames = rawGamesResponse.rows;
+    rawGames = mergeRawGames(rawGames);
```
ianbrown9475: Could you merge your changes from the other branch that moved this function to the serializer?

ianbrown9475 (changes requested):

ianbrown9475 (changes requested):
From my other comment https://github.com/osu-mist/poker-api/pull/10#pullrequestreview-260282136 could you fix the values of `cardSuit`? Feel free to merge the changes from the other branch after it's fixed

ianbrown9475 (approved):

PR #16, opened 2019-07-11
Feature/co1503 post members

nosnatef:


api/v1/db/oracledb/member-dao.js
```diff
@@ -43,27 +44,92 @@ const getMembers = async (query) => {
...
+      const memberCount = parseInt(rawMemberResponse.rows[0]['COUNT(1)'], 10);
+      return memberCount === memberIds.length;
+    }
+    return true;
+  } finally {
+    connection.close();
+  }
+};
+
+const postMember = async (body) => {
```
ianbrown9475, line 107: Could you add some documentation to this function?

ianbrown9475 (changes requested):
I queried `POST /members` with a valid member body and got a 201 back. Then, I repeated the request with the same body and got a 500: `Error: ORA-00001: unique constraint (MISTSTU2.SYS_C0032660) violated`. Could you look into this?
Could you also fix the merge conflicts?

api/v1/db/oracledb/member-dao.js
```diff
@@ -96,9 +97,64 @@ const validateMembers = async (memberIds) => {
   }
 };
 
+
+/**
+ * @summary Post a new member into the system.
+ * @function
+ * @param {Object} body The post body from request object.
+ * @returns {Object} The JSON resource of the new member created.
```
ianbrown9475: I think the return type should be `Promise<Object>` since the function is marked `async`

openapi.yaml
```diff
@@ -808,7 +810,7 @@ definitions:
         type: object
```
ianbrown9475, line 811: Could you update the `id` field of the member resource to be a string?

ianbrown9475 (changes requested):
I queried `POST /members` and the top-level self link was `https://localhost/v1/members/81` but it should be `https://localhost/v1/members`. Could you fix this?

ianbrown9475 (approved):

PR #21, opened 2019-07-17
Feature/co1521 delete player by gameid playerid

nosnatef:


api/v1/db/oracledb/player-dao.js
```diff
@@ -73,4 +73,38 @@ const getPlayerByGameIdAndPlayerId = async (id, pid) => {
...
 
-module.exports = { getPlayersByGameId, getPlayerByGameIdAndPlayerId };
+const cleanPlayerCardsByPlayerId = async (playerId) => {
+  const connection = await conn.getConnection();
+  try {
+    const sqlParams = [playerId];
+    const deleteSqlQuery = `
+    DELETE FROM PLAYER_CARDS WHERE PLAYER_ID = :playerId
+    `;
+    const result = await connection.execute(deleteSqlQuery, sqlParams, { autoCommit: true });
```
ianbrown9475, line 94: `cleanPlayerCardsByPlayerId()` is called by `deletePlayerByPlayerId`. Both SQL queries are executed in sequence with `autoCommit: true`. Here's a scenario where this could cause a problem:

1. Client requests `DELETE /games/1/players/1`
2. All rows in `PLAYER_CARDS` with `PLAYER_ID = 1` are deleted sucessfully and the transaction is committed
3. The `DELETE FROM PLAYERS` transaction fails due to a DB error
4. A 500 response is correctly returned to the client

The problem is that the resulting state is the player's cards being deleted without deleting the player. If a 500 response is returned to the client for their delete request, the expected state should be that nothing has changed, since API requests should be atomic (everything changes or nothing changes, no partial changes).

From our discussion, could you look into this?

ianbrown9475 (changes requested):

ianbrown9475 (approved):

PR #27, opened 2019-08-02
Feature/co1534 implement unit test member and game

nosnatef:


package.json
```diff
@@ -34,13 +38,16 @@
...
+    "gulp-if": "^3.0.0",
     "js-yaml": "^3.13.1",
     "jsonapi-serializer": "^3.5.6",
     "lodash": "^4.17.14",
     "moment": "^2.22.2",
     "oracledb": "^3.1.2",
+    "proxyquire": "^2.1.1",
     "query-string": "^6.2.0",
     "simple-git": "^1.96.0",
+    "sinon": "^7.3.2",
```
ianbrown9475: Could you move all the dependencies that won't be needed by the api to run using `gulp start` to `devDependencies` instead of `dependencies`?

tests/unit/test-util.js
```diff
@@ -0,0 +1,68 @@
...
+const testSingleResource = (serializedResource, resourceType, resourceId, nestedProps) => {
+  expect(serializedResource).to.containSubset(resourceSchema(resourceType,
+    resourceId,
+    nestedProps));
+};
+
+const testMultipleResources = (serializedResources) => {
+  expect(serializedResources).to.have.all.keys('data', 'links');
+  expect(serializedResources.data).to.be.an('array');
+  getDefinition('MemberResult');
```
ianbrown9475: What is `getDefinition()` doing here?

tests/unit/test-util.js
```diff
@@ -0,0 +1,68 @@
...
+  return schema;
+};
+
+const getDefinition = (def) => {
+  const result = openapi.definitions[def].properties.data.properties;
+  return result;
+};
+
+const testSingleResource = (serializedResource, resourceType, resourceId, nestedProps) => {
+  expect(serializedResource).to.containSubset(resourceSchema(resourceType,
```
ianbrown9475: Could you use `deep.equal()` instead of `containSubset()` here to make a stronger assertion?

tests/unit/test-util.js
```diff
@@ -0,0 +1,68 @@
...
+    resourceId,
+    nestedProps));
+};
+
+const testMultipleResources = (serializedResources) => {
+  expect(serializedResources).to.have.all.keys('data', 'links');
+  expect(serializedResources.data).to.be.an('array');
```
ianbrown9475, line 97: You should also test that each resource matches the resource schema here, similar to what you do in `testSingleResource`.

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,349 @@
...
+const { createConnStub } = require('./test-util');
+const {
+  fakeId,
+  fakeMemberPostBody,
+  fakeEmail,
+  fakeName,
+  fakeMemberPatchBody,
+} = require('./test-data');
+
+sinon.replace(config, 'get', () => ({ oracledb: {} }));
```
ianbrown9475: From our conversation, since you are calling `sinon.restore()` after each test case, this stub will be overridden after the first test case.

ianbrown9475: Could you use `sinon.stub()` here instead of `replace()`? The stub object provides more features so you can easily add more assertions later on if you need to

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,349 @@
...
+  describe('Test getMembers', () => {
+    const validTestCases = [
+      testCases.multiResult,
+      testCases.singleResult,
+      testCases.emptyResult,
+    ];
+    _.forEach(validTestCases, (testCase) => {
+      it(`getMembers should be fulfilled with ${testCase.description}`, () => {
+        createConnStub(testCase);
+        const membersSerializerStub = sinon.stub(membersSerializer, 'serializeMembers');
```
ianbrown9475: Could you rename this to something like `serializeMembersStub` since this is the method that you're stubbing? Could you also fix this in similar situations?

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,349 @@
...
+        membersSerializerStub.returnsArg(0);
+
+        membersDao = proxyquire(`${appRoot}/api/v1/db/oracledb/member-dao`,
+          { [`${appRoot}/api/v1/serializers/members-serializer`]: membersSerializerStub });
```
ianbrown9475: Since you're stubbing the `serializeMembers` method on `membersSerializer`, could you update the object that you're passing here to something like
```js
{
  [`${appRoot}/api/v1/serializers/members-serializer`]: {
    serializeMembers: serializeMembersStub,
  }
}
```
similar to [here](https://github.com/thlorenz/proxyquire/blob/5b88b0798f57767af63c89292faa0357181db8ea/examples/sinon/foo-tests.js#L42-L43)?
Could you also fix this in similar situations?

tests/unit/members-serializer.test.js
```diff
@@ -0,0 +1,45 @@
...
+      _.omit(rawMembers[0], id));
+  });
+
+  it('Should return an array of valid JSON objects that follow the OpenAPI specification', () => {
+    const { serializeMembers } = memberSerializer;
+
+    const serializedMembers = serializeMembers(rawMembers, testData.fakeMemberQuery);
+    testMultipleResources(serializedMembers);
+
+    for (let i = 0; i < serializedMembers.length; i += 1) {
```
ianbrown9475: I think you could use something like
```js
_.zipWith(serializedMembers, rawMembers, (a, b) => { ... })
```
where `a` is each member of `serializedMembers` and `b` is each member of `rawMembers` https://lodash.com/docs/#zipWith

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,349 @@
...
+    it('patchMember should be fulfilled with one row effected result and return true', () => {
+      createConnStub(testCases.rowEffectedOneResult);
+
+      const expectedResult = true;
+      const result = membersDao.patchMember(fakeId, fakeMemberPatchBody);
+      return result.should
+        .eventually.be.fulfilled
+        .and.deep.equal(expectedResult)
+        .then(() => {
+
```
ianbrown9475: There's a few cases like this where it looks like there's some missing logic. Could you either add something here or remove the `.then()`?

tests/unit/members-serializer.test.js
```diff
@@ -0,0 +1,45 @@
...
+  it('Should return a valid JSON object that follow the OpenAPI specification', () => {
+    const { serializeMember } = memberSerializer;
+    const serializedMember = serializeMember(rawMembers[0]);
+    testSingleResource(serializedMember,
+      resourceType,
+      rawMembers[0][id],
+      _.omit(rawMembers[0], id));
+  });
+
+  it('Should return an array of valid JSON objects that follow the OpenAPI specification', () => {
```
ianbrown9475: Could you separate these tests into `serializeMember` and `serializeMembers` categories?

ianbrown9475 (changes requested):
I tried running the tests without a config file and they failed. Could you look into this?

tests/unit/members-serializer.test.js
nosnatef: I don't see the reason to do that. The goal of this loop is not making a new array.

tsoliangwu0130: I guess the way @ianbrown9475 suggested didn't create a new array as well. His suggestion was actually a shorthand of what you exactly did here.

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
+const appRoot = require('app-root-path');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const chaiExclude = require('chai-exclude');
+const _ = require('lodash');
+const proxyquire = require('proxyquire');
+const sinon = require('sinon');
+const { createConnStub } = require('./test-util');
+const testData = require('./test-data');
```
Ruefa, line 10: Required files in the current directory should be in their own group. [Example](https://github.com/osu-mist/game-curator-api/blob/99e4eebe74872b5844139c4b17bb8adb8746e981/tests/unit/games-dao.test.js#L9)

tests/unit/test-util.js
```diff
@@ -0,0 +1,75 @@
...
+const testSingleResource = (serializedResource, resourceType, resourceId, nestedProps) => {
+  expect(serializedResource).to.deep.equal(resourceSchema(resourceType,
+    resourceId,
+    nestedProps));
+};
+
+const testMultipleResources = (serializedResources, rawResources, resourceType, resourceKey) => {
+  expect(serializedResources).to.have.all.keys('data', 'links');
+  expect(serializedResources.data).to.be.an('array');
+  _.zipWith(serializedResources.data, rawResources, (a, b) => {
```
tsoliangwu0130: You should name your functions/variables descriptively to avoid any ambiguity and confusion.

tsoliangwu0130, line 98: You should name your functions/variables descriptively to avoid any ambiguity and confusion. Also, can you add JSDoc strings to all of these functions to help others to understand your code easily?

tests/unit/test-data.js
```diff
@@ -0,0 +1,324 @@
...
+        round: 'river',
+        minimumBet: 10000,
+        maximumBet: 20000,
+        betPool: 1040000,
+        tableCards: [],
+      },
+ }],
+};
+const serializedGame = {
+  links: { self: '/v1/games/1' },
```
tsoliangwu0130, line 319: `/v1/games` was reused many times. You should store it somewhere and reuse it.

tests/unit/test-util.js
```diff
@@ -0,0 +1,75 @@
+const sinon = require('sinon');
+const { expect } = require('chai');
+const appRoot = require('app-root-path');
+const camelCase = require('camelcase');
+const _ = require('lodash');
+const config = require('config');
+
+sinon.stub(config, 'get').returns({ oracledb: {} });
```
tsoliangwu0130: This line should be after the require blocks.

tests/unit/test-util.js
```diff
@@ -0,0 +1,75 @@
...
+const conn = appRoot.require('api/v1/db/oracledb/connection');
+const { openapi } = appRoot.require('utils/load-openapi');
+const { fakeBaseUrl } = require('./test-data');
```
tsoliangwu0130: Base on our [style guide](https://github.com/osu-mist/style-guides/blob/master/JavaScript.md#require-statement-order), requires a local file should be put in front of the `appRoot` require.

package.json
```diff
@@ -34,6 +35,7 @@
     "express-basic-auth": "^1.1.7",
     "express-openapi": "^4.6.5",
     "express-winston": "^3.0.0",
+    "gulp-if": "^3.0.0",
```
tsoliangwu0130: This should be in `devDependencies`.

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
...
+
+      const result = gamesDao.postGame(fakeGamePostBody);
+      return result.should
+        .eventually.be.fulfilled
+        .and.deep.equal(expectedResult);
+    });
+  });
+
+  describe('Test validateGame', () => {
+    it('The function should be rejected when the database response does not contain COUNT(1) field', () => {
```
Ruefa: These 3 `COUNT(1)` tests can be done in a loop. Just make a list containing references to the different functions being tested.

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
+const appRoot = require('app-root-path');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const chaiExclude = require('chai-exclude');
+const _ = require('lodash');
+const proxyquire = require('proxyquire');
+const sinon = require('sinon');
+const { createConnStub } = require('./test-util');
```
tsoliangwu0130, line 9: A blank line is needed between NPM package require section and local file require section.

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
...
+const testData = require('./test-data');
+
+
```
tsoliangwu0130: One extra blank line here.

tsoliangwu0130 (changes requested):

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
...
+      });
+  });
+
+  afterEach(() => sinon.restore());
+
+  describe('Test getGames', () => {
+    const validTestCases = [
+      testCases.emptyResult,
+      testCases.multiResult,
+      testCases.singleResult,
```
tsoliangwu0130: You should destructure `testCases.emptyResult`, `testCases.multiResult`, `testCases.singleResult` and all other used test cases from `testCases` instead of indexing them every time when you need them.

tests/unit/games-serializer.test.js
```diff
@@ -0,0 +1,45 @@
...
+  } = testData;
+
+  afterEach(() => sinon.restore());
+
+  describe('Test mergeRawGames', () => {
+    it('Should merge rawgames properly', () => {
+      const result = gameSerializer.mergeRawGames(rawGames);
+      chai.expect(result).to.deep.equal(mergedRawGames);
+    });
+  });
```
tsoliangwu0130, line 27: I tried to run your unit tests but it failed:

```
  1) Test games-serializer
       Test mergeRawGames
         Should merge rawgames properly:
     TypeError: gameSerializer.mergeRawGames is not a function
      at Context.it (tests/unit/games-serializer.test.js:27:37)
```

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
...
+        const result = gamesDao.getGameById(fakeId);
+        return result.should
+          .eventually.be.fulfilled
+          .and.deep.equal(expectedResult);
+      });
+    });
+
+    it('getGameById should reject with multiple results', () => {
+      const testCase = testCases.multiGameResult;
+
```
tsoliangwu0130: Not necessary blank line.

tests/unit/games-serializer.test.js
```diff
@@ -0,0 +1,45 @@
+const appRoot = require('app-root-path');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const chaiSubset = require('chai-subset');
+const sinon = require('sinon');
+const testData = require('./test-data');
```
Ruefa: See comment in games-dao.test.js

tsoliangwu0130 (changes requested):

tests/unit/games-serializer.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+describe('Test games-serializer', () => {
+  const {
+    rawGames,
+    mergedRawGames,
+    serializedGames,
+    serializedGame,
+  } = testData;
+
+  afterEach(() => sinon.restore());
```
Ruefa: You aren't using sinon except for this line so it is not needed. You can remove the require() as well.

tests/unit/games-dao.test.js
```diff
@@ -0,0 +1,215 @@
...
+    it('Empty body should be fulfilled while the execute function from conn object is not called at all', () => {
+      const connStub = createConnStub();
+
+      const result = gamesDao.patchGame(fakeId, {});
+      return result.should
+        .eventually.be.fulfilled
+        .and.equal(true)
+        .then(() => {
+          sinon.assert.callCount(connStub.executeStub, 0);
```
tsoliangwu0130: You should use `sinon.assert.notCalled` here. Same with other places.

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
+const appRoot = require('app-root-path');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+const chaiExclude = require('chai-exclude');
+const _ = require('lodash');
+const sinon = require('sinon');
+const proxyquire = require('proxyquire');
+const { createConnStub } = require('./test-util');
```
Ruefa, line 9: See comment in games-dao.test.js

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
...
+            [`${appRoot}/api/v1/serializers/members-serializer`]: { serializeMembers: serializeMembersStub },
+          });
+
+        const expectedResult = testCase.data.rows;
+        const result = membersDao.getMembers();
+        return result.should
+          .eventually.be.fulfilled
+          .and.deep.equal(expectedResult)
+          .then(() => {
+            sinon.assert.callCount(serializeMembersStub, 1);
```
tsoliangwu0130: Should be `sinon.assert.calledOnce`, same with other places.

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
...
+const {
+  fakeId,
+  fakeMemberPostBody,
+  fakeEmail,
+  fakeName,
+  fakeMemberPatchBody,
+  testCases,
+} = require('./test-data');
+
+// sinon.replace(config, 'get', () => ({ oracledb: {} }));
```
Ruefa: Seems like this line is only needed in one test class. Remove the comment if it is not needed.

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
...
+        },
+        {
+          param: 'ABCHH',
+          result: 'ABCHH',
+        },
+        {
+          param: 'AbChC',
+          result: 'AB_CH_C',
+        },
+      ];
```
tsoliangwu0130: Should these be in `test-data.js` as well?

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
...
+        '0',
+        'false',
+        [],
+        [1, 2, 3],
+        {
+          a: 'b',
+        },
+        {},
+        () => {},
+      ];
```
tsoliangwu0130: Should these be in test-data.js as well?

tsoliangwu0130 (changes requested):

tests/unit/members-dao.test.js
```diff
@@ -0,0 +1,300 @@
...
+      const result = membersDao.validateMembers([fakeId]);
+      return result.should.eventually.be.rejectedWith(TypeError, error);
+    });
+  });
+
+  describe('Test hasDuplicateMemberId', () => {
+    it('Return false when empty or valid no dup array', () => {
+      const emptyResult = membersDao.hasDuplicateMemberId([]);
+      assert.strictEqual(emptyResult, false);
+      const validResult = membersDao.hasDuplicateMemberId([1, 4, 2, 3, 7, 6, 200]);
```
Ruefa: Should these be in test-data?

tests/unit/members-serializer.test.js
```diff
@@ -0,0 +1,48 @@
...
+const memberSerializer = appRoot.require('api/v1/serializers/members-serializer');
+
+chai.should();
+chai.use(chaiAsPromised);
+chai.use(chaiSubset);
+
+describe('Test memberSerializer', () => {
+  const { rawMembers } = testData;
+  const resourceType = 'member';
+  const id = 'MEMBER_ID';
```
Ruefa: `id` is not very descriptive. I would call it `memberId`

Ruefa (changes requested):
I had one failure when running tests
`1) Test games-serializer
       Test mergeRawGames
         Should merge rawgames properly:
     TypeError: gameSerializer.mergeRawGames is not a function
      at Context.it (tests/unit/games-serializer.test.js:27:37)`

Good work! I left some comments.

tests/unit/test-data.js
nosnatef: It's just the test data to be used to compared. I don't see reasons to refer to a variable in the test data.

tsoliangwu0130: Test data is also part of your code, it’s mainly for the good coding practice.

nosnatef: To me it's just aliasing instead of refactoring the code. It's just a dummy data and changing it would not affect the test at all.

tsoliangwu0130: It's totally your choice to decide how to write your code since no one else will need to maintain this API besides yourself.

My question is what's the reason not to do that even it's just an alias to you? How can you guarantee we won't need to change this string in the future? If that's the case, do we wanna find them all and replace them one by one? Since we're working as a group, a thing we need to always keep in mind is your code might need to be read and maintained by others in the future. If your code can be clearer and easier to maintain, why not doing it?

nosnatef: Sounds good but wouldn't that rule extends to all other strings in the test data? What's the boundary between raw data and reusable data?

tsoliangwu0130: I guess I'm not really sure what did you mean. If you were saying something like `diamonds`, `spades`, etc. those are constant and reused many times, then yes I'd agree to store them into constant strings to reuse. When we mentioned something like "this code was used many times", I think it's worthy to think about it more will there be a method to improve/combine/reuse the logic/function/variable to enhance your code. No matter it's in your test data or in the real code.
@ianbrown9475 since you're really good at writing clean code, can you also share some of your thought?

tests/unit/test-util.js
```diff
@@ -5,11 +5,18 @@ const camelCase = require('camelcase');
...
+
 const conn = appRoot.require('api/v1/db/oracledb/connection');
 const { openapi } = appRoot.require('utils/load-openapi');
-const { fakeBaseUrl } = require('./test-data');
 
+sinon.stub(config, 'get').returns({ oracledb: {} });
+/**
+ * @summary Create the stub for connection object that is useful for controlled testing environment.
+ * @function
+ * @param {Object} testCase The testcase that becomes the return value of connection.execute in the stub.
```
Ruefa: We are transitioning to lowercase o on `object`. Can you make this change in your other classes?

tests/unit/test-util.js
```diff
@@ -22,6 +29,14 @@ const createConnStub = (testCase) => {
   };
 };
 
+/**
+ * @summary Transoform the rawData into serializedData.
```
Ruefa: You can remove the `@summary` tag and the `@function` line. [Example](https://github.com/osu-mist/game-curator-api/blob/4929a92be94da94b9fb2084382360c9832da033b/tests/unit/test-helpers.js#L38-L44).

tests/unit/games-dao.test.js
```diff
@@ -120,43 +126,32 @@ describe('Test games-dao', () => {
...
-
-  describe('Test isMemberInGame', () => {
-    it('The function should be rejected when the database response does not contain COUNT(1) field', () => {
-      const testCase = testCases.emptyResult;
-      const error = 'Cannot read property \'COUNT(1)\' of undefined';
-
-      createConnStub(testCase);
-      const result = gamesDao.isMemberInGame(fakeId);
-      return result.should.eventually.be.rejectedWith(TypeError, error);
+  it('', () => {
```
Ruefa: I don't think this `it('', () => {` needs to be here

tests/unit/test-data.js
ianbrown9475: I think @tsoliangwu0130 makes a really good point that the test data, even though it's just dummy data, is still code that we're committing to source control and being used by the unit tests. Even though it may seem like random data created for a simple purpose, we need to treat it like actual code, since it's still going to be maintained. For our other APIs, this means that other developers will have to eventually go back, look at the code, and modify it when there's a code/design change.

I would suggest writing these so that another developer can easily open this file for the first time and make changes as easily as possible. Variables that have values like `v1` or `/v1/games` should probably go at the very top of this file, with some documentation on them. This way, if the api version is bumped to `v2` for example, developers can open this file, see where this `v1` variable is used, and replace it once with `v2`. Then, they don't have to find all instances of `v1` and replace it manually.

For values like `diamonds`, `spades`, `A`, `2`, `Q`, `blind`, ...; these could be stored in a variable as well. Ideally, you would load these enum values from your openapi document, but that may not be necessary for this learning project. @nosnatef as far as when to refactor these values, I think it's important to think about how easy the test data would be to change if there was a design change.

tests/unit/games-dao.test.js
```diff
@@ -127,11 +127,11 @@ describe('Test games-dao', () => {
...
     const countOneFunctions = [
-      gamesDao.validateGame,
-      gamesDao.getGamesByMemberId,
-      gamesDao.isMemberInGame,
+      'validateGame',
+      'getGamesByMemberId',
+      'isMemberInGame',
     ];
     _.forEach(countOneFunctions, (fun) => {
       describe(`Test ${fun.name}`, () => {
```
Ruefa: I believe this should just be `fun` instead of `${fun.name}` now that you changed `countOneFunctions`

